PR #5, opened 2016-10-28
Fixed the rotated method of all objects

remi-pr:
The rotated method of all objects was not using the origin argument but always self.point(0.5) even though it is all correctly handled in the rotate function they are all calling


remi-pr:
Thanks for merging this in. Found this because this library is a great tool, happy to be able to contribute a little bit.

mathandy:
@remi-pr 
No problem and thanks!  That's cool to hear you like it.

PR #7, opened 2017-02-04
unset svgwrite debug mode

derVedro:
Debug flag is set by default so that svgwrite do some validation stuff.
If you pass paths with many segments or generally a lot of paths svgwrite
takes a lot of time for validation. The paths that svgpathtools fed are
all of correct syntax and bad attributes are ignored by almost all programs
that handled with SVG. So, setting svgwrite debug to False is good choice.

### DIFF
--- a/svgpathtools/paths2svg.py
+++ b/svgpathtools/paths2svg.py
@@ -276,9 +276,9 @@ def disvg(paths=None, colors=None,
 
     # Create an SVG file
     if svg_attributes:
-        dwg = Drawing(filename=filename, **svg_attributes)
+        dwg = Drawing(filename=filename, debug=False, **svg_attributes)
     else:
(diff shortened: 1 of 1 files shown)


mathandy:
Hi derVedro, thanks for pointing this out!  
I don't know what the `debug` flag in `svgwrite.Drawing` does and I'm not sure why the makers of `svgwrite` set this flag to `True` by default.  Without that knowledge I'm too nervous to merge in your requested default overriding of their default.  Especially since...

**A simple (though not obvious) workaround also exists:**

    from svgpathtools import *
    wsvg(Path(Line(0, 100+100j)), svg_attributes={'debug': False})

From the one `%timeit` test I did, this does appear to provide a significant speed increase.  I just made a note of this in the `disvg` docstring.

PR #8, opened 2017-02-20
fix docstring

derVedro:


derVedro:
It was not clear to me that straight lines are a beziers too. sorry, I close it.

PR #11, opened 2017-02-28
xrange & fix formula

derVedro:
updated for python3, found that else branch was broken, fixed.


mathandy:
Thanks again @derVedro.  This actually isn't a complete fix.  Then coefficients for higher order (degree > 3)  polynomials needed to be reversed also.  I'm surprised you have a use for higher order Bezier curves actually.  Anyways it's fixed now.  BTW, I'll rollout a "futurized" update soon to (hopefully) fix any remaining Python 3 issues.

PR #13, opened 2017-03-15
polyline2pathd was already adding the z command. Hence, adding the z …

jpcofr:
…twice caused closed shapes to be considered open shapes.

Change-Id: I36977ca9dc9611ffad2309e485fbd72e44fe3ff6

### DIFF
--- a/svgpathtools/svg2paths.py
+++ b/svgpathtools/svg2paths.py
@@ -87,7 +87,7 @@ def svg2paths(svg_file_location,
     # path strings, add to list
     if convert_polygons_to_paths:
         pgons = [dom2dict(el) for el in doc.getElementsByTagName('polygon')]
-        d_strings += [polyline2pathd(pg['points']) + 'z' for pg in pgons]
+        d_strings += [polyline2pathd(pg['points']) for pg in pgons]
         attribute_dictionary_list += pgons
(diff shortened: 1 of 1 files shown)


mathandy:
@jpcofr Can you give me an example of what you mean?  The following example seems to work as expected.

```
<svg height="210" width="500">
  <polygon points="0,0 0,100 100,100" style="stroke:purple;stroke-width:1"/>
</svg>
```

jpcofr:
With the following example, `polyline2pathd(...)` decided that the path is closed (lines 21-24) and added a trailing 'z' (lines 29-39).   

` <polygon points="438,-16 438,-884 1218,-884 1218,-16 438,-16"/>`

Then, on line 83, a new 'z' was added, so the final string caused the new Path to appear as closed when checking the property  `@closed.setter` so the error "End does not coincide with a segment start." is raised when trying to find a bounding box.

mathandy:
@jpcofr 
You're correct, svgpathtools was choking on redundantly closed polygons.  This is fixed now.

I decided the desired behavior should be to have a k-pointed SVG-Polygon element become a k-Line Path object.  This way information encoded as a k-pointed polygon is preserved in the transformation to a path object.

PR #25, opened 2017-07-19
now disvg and wsvg pass `debug=False` into svgwrite.Drawing() by default

mathandy:
@derVedro you asked me to do this a while back.  Have you noticed any hazard with this since or do you still agree with the change?


derVedro:
Nope, in any of my cases there were valid svgs in outcome. I am glad you change the flag to false.

PR #31, opened 2017-11-10
Arc parsing round-off fix

andersgb:
Add a unittest to demonstrate the issue, and introduce a fix by clamping `acos` input to `[-1, 1]` with `round()`


mathandy:
Thanks @andersgb

PR #34, opened 2017-12-18
README: fix a type error in Offseting Paths example

SebKuzminsky:
This PR addresses a couple of minor problems with the Offsetting Paths example in the README.


mathandy:
Thanks @SebKuzminsky

PR #36, opened 2017-12-26
Improve Arc/Line intersections

SebKuzminsky:
I've found two problems with Arc.intersect():
1. Arc.intersect() returns all intersections between the specified Line and the *complete ellipse* implied by the Arc, including intersection points that do not lie between the `start` and `end` of the Arc.
2. Arc.intersect() returns parameterized `t` values that do not fall in the expected [0.0, 1.0] range.  When evaluated by Arc.point(), these out-of-range `t` values *do* specify cartesian/complex points that correspond with the actual intersection points (except fot the spurious `t` values returned that correspond with intersections of the ellipse outside the `start` to `end` range).

This PR fixes the first of the two problems by filtering out from the list of proposed intersection points the ones that do not lie between `start` and `end` on the Arc.

It also and adds unit tests for *both* problems, so it adds unit tests that fail, which I'm not sure if you want.  I'd be happy to change the branch so that it only tests for the problem that it fixes, so that the net number of failing unit tests doesn't go up.  (I'd put the tests for the second problem on a different branch.)  Just let me know what you want.


SebKuzminsky:
@mathandy did you click Approve when you intended to click Merge, or is the merge waiting for something?

mathandy:
Checked through your test cases.  Didn't go through your fixes yet.

SebKuzminsky:
I added another failing test that reproduces an arc/arc intersection problem I ran into.  The Arcs look like this (note that the big arc is open at the bottom):

arcs

SebKuzminsky:
I'm closing this PR because I have a better arc/line intersection fix coming in a different PR, and because I ill-advisedly added those arc/arc intersection tests that confuse things.

I'll deal with arc/arc intersections later.

PR #38, opened 2017-12-31
Arc intersection bbox

SebKuzminsky:
This PR adds a unit test for a bug in Arc/Line intersections, and then patches Arc.intersect() to work around this particular situation.


SebKuzminsky:
I'm closing this PR because I've got a better arc/line intersection fix coming in another PR.

Issue #41, opened 2018-01-07
Bug in Line.intersect(Line)


SebKuzminsky:
I found a bug where two obviously non-intersecting Lines still report an intersection using Line.intersect().  The bug is captured in this commit: SebKuzminsky/svgpathtools@847b270b 
It is highly sensitive to the precise starting and ending points of the two lines - if you round the values down to three significant places (like Line.\_\_repr\_\_() does) then the bug does not trigger.

PR #42, opened 2018-01-07
Fix Line.intersect(Line) (bug #41)

SebKuzminsky:
This PR adds a test that demonstrates #41, then fixes it.


SebKuzminsky:
Oops, I just ran in to a bug that I introduced in my original attempt at a fix.  It's explained and fixed in the commits I just added to this PR.

PR #48, opened 2018-03-04
use pretty printing in svgwrite.

inconvergent:
Hi

I noticed that svgwrite supports pretty printing by default. 

See: https://github.com/mozman/svgwrite/blob/e606a0c91d4f2e85aed4250aaccf4a05ea1b316f/svgwrite/drawing.py#L87 

I don't know if there is a specific reason why this was not used (maybe you are supporting older versions of svgwrite?), but I thought I'd submit a PR just in case. 

Feel free to ignore this, or tell me what else needs to be done to accept it. I've tested it locally, but might very well have made a mistake. 

Also, if this is an acceptable change: Maybe we should make pretty printing optional?

Anyway, let me know. 

-A

### DIFF
--- a/build/lib/svgpathtools/paths2svg.py
+++ b/build/lib/svgpathtools/paths2svg.py
@@ -5,7 +5,6 @@ segments."""
 from __future__ import division, absolute_import, print_function
 from math import ceil
 from os import getcwd, path as os_path, makedirs
-from xml.dom.minidom import parse as md_xml_parse
 from svgwrite import Drawing, text as txt
 from time import time
 from warnings import warn
@@ -153,7 +152,7 @@ def disvg(paths=None, colors=None,
 
         :param svg_attributes - a dictionary of attributes for output svg.
         Note 1: This will override any other conflicting settings.
-        Note 2: Setting `svg_attributes={'debug': False}` may result in a 
+        Note 2: Setting `svg_attributes={'debug': False}` may result in a
         significant increase in speed.
 
     NOTES:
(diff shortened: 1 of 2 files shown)


jf---:
That looks cool, but there's 2 file (write) operations, right? Perhaps writing to a file like buffer ( `io.StringIO` ) and prettifying that could allow avoiding writing the same file twice

inconvergent:
if we use my suggested change then this is already taken care of courtesy of svgwrite. see: https://github.com/mozman/svgwrite/blob/e606a0c91d4f2e85aed4250aaccf4a05ea1b316f/svgwrite/drawing.py#L110

inconvergent:
(deleted my initial response because it was wrong) it seems like you are correct. there is a previous write that i didn't really notice in the fog of debugging. i suggest we put this on hold and i can see if i can rewrite it.

jf---:
>  i will stop now

Please dont, btw fan of your (art)work

mathandy:
Hey @inconvergent and @jf---, thanks for your team effort here :)

I don't remember why I didn't use `svgwrite`'s pretty-print functionality.  

Is there an advantage to using it?
Is this pull-request still requested?

inconvergent:
@mathandy the reason i started looking at this was that i was having performance issues. it turned out not to be because of this in particular. (there were some debug-like flags that were enabled somewhere. and for various reasons i ended up forking the whole thing and adapting it to my use.). In short: i don't particularly need this any more.

mathandy:
OK thanks @inconvergent.
In that case, I'll close this.

PR #51, opened 2018-04-13
Issue16

mathandy:


svgpathtools/svg2paths.py
```diff
@@ -144,30 +174,139 @@ def dom2dict(element):
...
+            value_str = trafo_sub_str.split('(')[1]
+            values = list(map(float, value_str.split(',')))
+            if 'translate' in trafo_sub_str:
+                x = values[0]
+                y = values[1] if (len(values) > 1) else 0.
+                trafo_matrix = np.dot(trafo_matrix, np.array(
+                    [1., 0., x, 0., 1., y, 0., 0., 1.]).reshape((3, 3)))
+            elif 'scale' in trafo_sub_str:
+                x = values[0]
+                y = values[1] if (len(values) > 1) else 0.
```
mxgrey: I just thought I'd point out, according to the [W3C spec](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform), `scale` should assume that the y-scaling is equal to the x-scaling if the y-scaling is not explicitly specified. In this code, it sets the y-scaling to `0` when it's not specified, which would shrink paths down to being purely horizontal.

I'm working on an alternative PR for this, but anyone who wants to use this current PR for reference should be careful here.

svgpathtools/svg2paths.py
```diff
@@ -144,30 +174,139 @@ def dom2dict(element):
...
+                trafo_matrix = np.dot(trafo_matrix,
+                                      np.array([x, 0., 0., 0., y, 0., 0., 0.,
+                                                1.]).reshape((3, 3)))
+            elif 'rotate' in trafo_sub_str:
+                a = values[0] * np.pi / 180.
+                x = values[1] if (len(values) > 1) else 0.
+                y = values[2] if (len(values) > 2) else 0.
```
mxgrey: For `rotate`, according to the spec, either x and y should **both** be specified, or **neither** of them should be specified. If x is specified without y, then the SVG is malformed.

mathandy:
Looks like we have a solution to this now thanks the new Document class pushed recently by @mxgrey !
I haven't done much testing, but I'm impressed -- it looks like really good work.

I'll leave this issue open for a bit to encourage input on any minor changes that should be implemented before posting this on the PyPI -- it'll be tagged `svgpathtools` v2.0 when I do post it.

The one significant change I'm likely to implement is to have `flatten_all_paths` return a list of `Path` objects -- the `ElementTree` element and transform can be stored as attributes.  Do you see any problems with this @mxgrey ?

Two minor changes so far:
* all `np.matrix()` calls were replaced with `np.array()` -- `numpy` is deprecating `matrix` objects.
* the `save()` method now requires a filename.

PR #53, opened 2018-05-19
After Z, next subpath starts at same position

### LINKED ISSUE #52: subpaths do not need to start with M/m
For example `m0 0h50v50zv20h20z` right now causes an error:
>Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\python36\lib\site-packages\svgpathtools-1.3.2-py3.6.egg\svgpathtools\parser.py", line 81, in parse_path
  File "C:\python36\lib\site-packages\svgpathtools-1.3.2-py3.6.egg\svgpathtools\path.py", line 1883, in closed
>ValueError: End does not coincide with a segment start.

But this is ok

>[If a "closepath" is followed immediately by any other command, then the next subpath starts at the same initial point as the current subpath.](https://www.w3.org/TR/SVG/paths.html#PathDataClosePathCommand)

mdejean:
Fixes #52

### DIFF
--- a/svgpathtools/parser.py
+++ b/svgpathtools/parser.py
@@ -80,7 +80,7 @@ def parse_path(pathdef, current_pos=0j):
                 segments.append(Line(current_pos, start_pos))
             segments.closed = True
             current_pos = start_pos
-            command = None  # You can't have implicit commands after closing.
+            command = None
 
         elif command == 'L':
             x = elements.pop()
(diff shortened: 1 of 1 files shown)


mdejean:
The [spec](https://www.w3.org/TR/SVG/paths.html#PathData) does not mention "implicit commands", rather the moveto, line, curve, and arc commands can have multiple arguments (and for moveto they make lines). Z takes no arguments.

mathandy:
Hi @mdejean I'm not sure what you mean -- the spec (the same linked to, in section 8.3.2) _does_ mention "implicit" commands:

> "If a moveto is followed by multiple pairs of coordinates, the subsequent pairs are treated as **implicit lineto commands**. Hence, **implicit lineto commands** will be relative if the moveto is relative, and absolute if the moveto is absolute."

PR #54, opened 2018-05-22
add scale() for curves, and scaled() for paths

elenzil:


mathandy:
Hey @elenzil nice code!  I was about to merge this, but I realized that the behavior of your scale(d) function/methods doesn't match the behavior of the of the SVG-Path scale transform attribute.

[Official reference](https://www.w3.org/TR/SVG/coords.html#TransformAttribute)
[perhaps more clear reference](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform)

I see a few options (any of which sound fine to me):
1. change the behavior to match the official scale behavior.
2. change the name of scale (e.g. `dilate` might be a good choice)
3. both (1) and (2)

What do you think?

elenzil:
hi @mathandy !

thanks for looking at this so fast, sorry i missed your comment until now.
thanks also for the style changes.
i think you're right.
i didn't feel up to writing the unit tests etc for non-uniform scale,
but it would be good to change the name to indicate.
`dilate` is an interesting word, but i've never encountered it before in transformation language.
i'd probably go for `scale_uniform` so that users who are looking for `scale` will be likely to find it.

i'll make that change.

also fwiw, this is my first cross-fork pull-request, so apologies if i mess up on the mechanics there.

mathandy:
@elenzil, yeah, "dilation" doesn't seem to be used much outside math, and it's a bit confusing of a term to begin with.  That said "scale_uniform" is not a common term as far as I know --  though I do like your point about it being found through searches -- maybe we should go with "scalar_transform".
Or maybe we should go with "scale" and simply implement the full functionality -- it seems pretty trivial to do so, the only part I'm not sure of is how Arc elements behave under scale transforms.

Thoughts?

mathandy:
@elenzil So how about that?

elenzil:
yeah, Arcs seem like a trick. i'm confident that any non-uniform scaling of an arc can be represented by another arc, but my guess is that at least three parameters of arc would need to change, one of them being rotation.

i'll leave the choice to you. imo 'scale_uniform' is a reasonable name, and worth the effort so far.

Issue #47, opened 2018-02-13
polyline2pathd(): Conversion from polyline could not handle strings with multiple spaces.


ctc53:
Said there is a string in SVG:<b> \<polyline points="10,&nbsp;&nbsp;20 30,40"/\></b>  #two spaces before 20
polyline2pathd() would return the conversion as u'M10 L20L30 40' rather than u'M10 20L30 40'.

DmitryMilk:
Furthermore, polyline2pathd and polygon2pathd are not compliant with EBNF-specification of attribute 'points' in ‘polyline’ and ‘polygon’ elements:

https://www.w3.org/TR/SVG11/shapes.html#PointsBNF

According to this specification, all the cases listed below are valid and equivalent to "10,20 30,40":

- "10,20,30,40" (pairs also can be delimited by comma. Violation of this case I encountered with)
- "10 20&nbsp;&nbsp;&nbsp;,&nbsp;&nbsp;&nbsp;30 ,
40" (any number of whitespaces/tabulations/LFs/CRs allowed  around optional comma)
- "10-20 300e-1-4000E-2" (negative Y can follows immediately after X. Exponential form of floating point must be taken in account and processed properly)

PR #56, opened 2018-05-27
Polyline/polygon parsing adjusted to specification

DmitryMilk:
Dealing with #47 
Adjusted to https://www.w3.org/TR/SVG11/shapes.html#PointsBNF

### DIFF
--- a/svgpathtools/svg2paths.py
+++ b/svgpathtools/svg2paths.py
@@ -5,10 +5,16 @@ The main tool being the svg2paths() function."""
 from __future__ import division, absolute_import, print_function
 from xml.dom.minidom import parse
 from os import path as os_path, getcwd
+import re
 
 # Internal dependencies
 from .parser import parse_path
 
+COORD_PAIR_TMPLT = re.compile(
+    r'([\+-]?\d*[\.\d]\d*[eE][\+-]?\d+|[\+-]?\d*[\.\d]\d*)' +
+    r'(?:\s*,\s*|\s+|(?=-))' +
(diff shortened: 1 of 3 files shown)


svgpathtools/svg2paths.py
```diff
@@ -37,19 +43,21 @@ def ellipse2pathd(ellipse):
...
+    # This check ensures that an n-point polygon is converted to an n-Line path.
+    if is_polygon and closed:
+        points.append(points[0])
 
-    d = 'M' + points.pop(0).replace(',', ' ')
-    for p in points:
-        d += 'L' + p.replace(',', ' ')
-    if closed:
+    d = 'M' + 'L'.join('{0} {1}'.format(x,y) for x,y in points)
+    if is_polygon or closed:
```
mathandy: The `or closed` seems a bit dangerous -- the "z" being redundant in this case.  Any argument for why to leave this?

mathandy:
Thanks @DmitryMilk ! The code looks good.  I made one in-line comment/review.

svgpathtools/svg2paths.py
DmitryMilk: I did not change the logic of parent commit:
- in polygon2paths 'z' was appended in any case (along with appending of an additional segment),
- in polyline2paths 'z' was appended if the ends of the path were equal.
This condition remains that logic.

mathandy: OK, perfect! Thanks for the explanation :)

PR #65, opened 2018-08-23
Update Path.area to work with Arc segments

SebKuzminsky:
It now approximates Arc segments as a sequence of chords, with configurable length.

This PR also includes a couple of tests that verify that Path.area() is broken before this PR, and works after this PR.


mathandy:
Thanks again @SebKuzminsky!

PR #66, opened 2018-08-28
Basic Svg_Io_Sax

tatarize:
Demonstration of the load as flatten capabilities discussed in mathandy/svgpathtools#16 shows the other method svg parsing, in the case that speed matters.


mathandy:
Looks good, thanks @tatarize.  I don't see any reason not to merge this in -- haven't tested it myself to be honest.  If you want people to experiment with it and improve on it you might consider adding usage examples to the module docstring -- otherwise (unless popular demand is that this becomes the main IO method) I might move it to a `contrib` submodule in the future.  That's not to say it's not superior -- my mind is just other places these days, so I'm leaving a lot up to popular demand with `svgpathtools`.

Issue #68, opened 2018-09-06
Decouple the Parser from the Path


tatarize:
As is the parser for the SVG paths instances the Path object and the Line objects and the curve objects. It would be better if it didn't do that. Passing the path object into the parse_path command and as it parses the commands it would call a command on path that add_line() or add_curve() or start(), or calling these on a different class that simple delegated to segments.append(Line(start_pos,end_pos)) or whatever.

It would make that code rather independently useful, as anybody could cook up a parser object and throw it in there, even if they weren't going to actually use the rest of the path tools as they were intended. The code has some reasonable svg parse capabilities and those could be broken off and decoupled more effectively from the math commands and logic in the primary classes.

This could even be taken to the somewhat logical extreme of branching that project off completely and writing an `svgread` module, which would be completely ambiguous as to where the data ends up. Just including the parsing tools as such loading the data in either a proper dom or the various other methods of loading and extracting that data. Including the svg path parsing. Which would basically yield proper calls parse that data.  As is the intersection between the parser bits, the svgpathtools and svgwrite all seems kind of sloppy.

tatarize:
See the relevant pull request. It also adds in the svg 2.0 modified z command. But, it would allow the parser.py file to be parsed off. And used mostly independent of the underlying classes being utilized. So most of the parsing would be decoupled from the relevant classes. Without the coupling between the parsing and the geometric classes that are the core of this project, underlying point of the classes could be better displayed. Even to the point of spinning off the parsing as being completely unrelated to the core functionality here.

tatarize:
Removed most of the core code off parser and put into svg_parser, and added svg_color parsing. There's a bunch of potential pitfalls people can hit trying to do that themselves. It's pretty solid as a standalone parsing file now. It could still eat up some parts of document and svg_to_paths. And some of the overlap in svgwrite and document too.

tatarize:
Many of these changes were integrated into my svgelements project with the fuller more robust svg parsing. The division of the parser and path objects however went the other way since there's really parsing svg is of little utility without the path objects.  I added a bunch of other things like Matrix and Color and Length elements to match the SVG/CSS usages and rather robustly parse everything. 

But, this ended up being a mistake, it's actually far better off to make the parser an invisible part of the Path object itself.

PR #69, opened 2018-09-06
Decoupled Pathd parser from Path objects.

tatarize:
With regard to #68, this gives a clear example of decoupled path parser from the path object. Note that both the SvgMatrix and the PathParser objects can be removed from the parser class and then the parser class will require `re` as the sum of the imports. The PathTokens class is even subclassed out from the SVGPathTokens class so if you want to deserialize generic svg-path-like data into a set of commands or subclass the SVGPathTokens class to parse a couple extra elements it otherwise wouldn't.

I also, added in the SVG 2.0 spec format variant of the 'z' command. Namely that any missing location values can be truncated to z. So "m 20,20 40,20, 40,40 z" is a valid path, among other things. You can curve_to a location z. Which is back where you started.


mathandy:
Hey @tatarize , you've been a long time and very helpful contributor to svgpathtools.

I just wanted to thank you and let you know I'll be accepting your pull requests by default going forward.  I'll also accept any pull requests that you approve.

FYI I told @SebKuzminsky the same.

Thanks very much for all your contributions over the years,
Andy

tatarize:
I forget how old this one is.  It might have some problematic conflicts.

My own code via https://github.com/meerk40t/svgelements actually goes the other way. I not only don't decouple to two I put them together much more tightly such that `Path("M5,5v2")` etc is a valid path. It just parses the sucker from the declaration. Which turns out to be a really nice way to do that, and is extremely intuitive.

tatarize:
Yeah. While I like some of this code, the truth is the path parser really needs to be a much more coherent part of the path object itself, rather than decoupled completely it should be integrated completely such that you never need the parser since you can simply call `Path(pathd_str)`

PR #79, opened 2018-11-16
Implement affine transform for paths and bezier segments

ArcturusB:
This PR proposes an implementation of affine transformation for `Path` and bezier segment objects. This is done by adding the function `transform()` to `svgpathtools/path.py`, which allows to transform the input objects with an arbitrary affine transformation matrix. 

I may submit a later PR that adds support for `Arc` objects, but it still requires some work.


ArcturusB:
Sorry, this was some old code I had on my computer. I didn’t realize the functionality had been implemented in the mean time.

PR #85, opened 2018-12-12
fix polyline and polygon conversion

NataliaTs:
When svg file that contains polylines or polygons  is converted to paths using Document method flatten_all_paths() an exception occurring in polyline2pathd() method:
TypeError: expected string or bytes-like object

Changing polyline2pathd() method to receive polyline instead of polyline_d makes both methods: flatten_all_paths() and svg2paths() working correctly when converting polylines and polygons.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ import codecs
 import os
 
 
-VERSION = '1.3.4'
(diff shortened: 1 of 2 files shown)


abey79:
I ran into this issue as well.

PR #86, opened 2019-01-04
Fix tuple assignment

skef:
Fix list-style assignment to tuple on line 309 of paths2svg.py

### DIFF
--- a/svgpathtools/paths2svg.py
+++ b/svgpathtools/paths2svg.py
@@ -306,8 +306,8 @@ def disvg(paths=None, colors=None,
 
     # Create an SVG file
     if svg_attributes is not None:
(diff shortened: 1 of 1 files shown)


mathandy:
Hey @skef, is there an issue with the way it was?
Why the need for this pull-request?

skef:
`dimensions` is generally a reference to a tuple (see, for example, lines 246 and 248 of the modified file). Python tuples (unlike lists) are immutable -- to modify a reference to a tuple you need to supply a new tuple. If you test the code with a tuple already assigned to `dimensions` you'll see that it fails. 

I submitted the pull request because I ran into the issue in practice. (If you're curious I used your library to make the outlne SVGs for this review: https://github.com/libertinus-fonts/libertinus/pull/251#issuecomment-451648353 )

PR #101, opened 2019-11-20
Fixed crash in line2pathd()

abey79:
`flatten_all_paths()` crashes with some SVG file in `line2pathd()`, which appears to be incorrectly implemented (attributes cannot be accessed with sub-scripting).

@mxgrey: it appears this code is from you, please comment if I'm missing something.

Edit: [here](https://cdn.discordapp.com/attachments/499297705588424715/645611959378771968/crosshatch_scribble_20191117-131012.svg) is an example of SVG which crashes

### DIFF
--- a/svgpathtools/svg_to_paths.py
+++ b/svgpathtools/svg_to_paths.py
@@ -93,8 +93,8 @@ def rect2pathd(rect):
 
 def line2pathd(l):
     return (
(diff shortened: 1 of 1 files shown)


abey79:
I further fixed my patch with support for implicit attributes (`x1`, `x2`, etc. must default to 0 if missing).

PR #114, opened 2020-06-25
Move Path Parsing into Path.

tatarize:
The path and parser are intimately linked, but still used separately.  The utility and usability of this goes up significantly if we are permitted to define paths with contexts like `Path('M0,0v50h50v-50z')`.

* Test coverage is added to prove the parser code and Path(path_d, current_pos) is identical
* No breaking changes.
* The parse_path function in `parser.py` is reduced to a stub for legacy reasons. 

Defining a path by using the Path's constructor is much more intuitive.


mathandy:
Looks good, thanks @tatarize !

Issue #103, opened 2020-02-22
Convert arc to cubic bezier


qo4on:
How to convert arc to cubic bezier with this tool?

tatarize:
There doesn't seem to be such a function but they usually look something like this (taken from https://github.com/meerk40t/svgelements )

    def as_cubic_curves(self):
        sweep_limit = tau / 12
        arc_required = int(ceil(abs(self.sweep) / sweep_limit))
        if arc_required == 0:
            return
        slice = self.sweep / float(arc_required)

        theta = self.get_rotation()
        rx = self.rx
        ry = self.ry
        p_start = self.start
        current_t = self.get_start_t()
        x0 = self.center[0]
        y0 = self.center[1]
        cos_theta = cos(theta)
        sin_theta = sin(theta)

        for i in range(0, arc_required):
            next_t = current_t + slice

            alpha = sin(slice) * (sqrt(4 + 3 * pow(tan((slice) / 2.0), 2)) - 1) / 3.0

            cos_start_t = cos(current_t)
            sin_start_t = sin(current_t)

            ePrimen1x = -rx * cos_theta * sin_start_t - ry * sin_theta * cos_start_t
            ePrimen1y = -rx * sin_theta * sin_start_t + ry * cos_theta * cos_start_t

            cos_end_t = cos(next_t)
            sin_end_t = sin(next_t)

            p2En2x = x0 + rx * cos_end_t * cos_theta - ry * sin_end_t * sin_theta
            p2En2y = y0 + rx * cos_end_t * sin_theta + ry * sin_end_t * cos_theta
            p_end = (p2En2x, p2En2y)
            if i == arc_required - 1:
                p_end = self.end

            ePrimen2x = -rx * cos_theta * sin_end_t - ry * sin_theta * cos_end_t
            ePrimen2y = -rx * sin_theta * sin_end_t + ry * cos_theta * cos_end_t

            p_c1 = (p_start[0] + alpha * ePrimen1x, p_start[1] + alpha * ePrimen1y)
            p_c2 = (p_end[0] - alpha * ePrimen2x, p_end[1] - alpha * ePrimen2y)

            yield CubicBezier(p_start, p_c1, p_c2, p_end)
            p_start = Point(p_end)
            current_t = next_t

Not sure how compatible the code is anymore.  But basically you calculate the control points that best approximate arc in some number of cubics.

PR #116, opened 2020-07-29
converting arcs to line segments is very inefficient, instead convert to cubic beziers

njhurst:
shamelessly based on https://github.com/fontello/svgpath/blob/master/lib/a2c.js https://github.com/colinmeinke/svg-arc-to-cubic-bezier

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,2 @@
+ [pytest]
+ addopts=-n12
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,0 @@
- [pytest]
- addopts=-n12
--- a/svgpathtools/arctobezier.py
+++ b/svgpathtools/arctobezier.py
@@ -1,0 +1,162 @@
+import math
+
+TAU = math.pi * 2
+
+def mapToEllipse(P, rx, ry, cosphi, sinphi, centerx, centery):
+    x, y = P
+    x *= rx
+    y *= ry
+
+    xp = cosphi * x - sinphi * y
+    yp = sinphi * x + cosphi * y
+
+    return (xp + centerx, yp + centery)
+
+def approxUnitArc(ang1, ang2):
+    # If 90 degree circular arc, use a ant
+    # as derived from http:#spencermortensen.com/articles/bezier-circle
+    a = (0.551915024494 if (ang2 == 1.5707963267948966)
+        else (-0.551915024494 if (ang2 == -1.5707963267948966)
+            else 4 / 3 * math.tan(ang2 / 4)))
+
+    x1 = math.cos(ang1)
+    y1 = math.sin(ang1)
+    x2 = math.cos(ang1 + ang2)
+    y2 = math.sin(ang1 + ang2)
+
+    return [
+        (
+            x1 - y1 * a,
+            y1 + x1 * a
+        ),
+        (
+            x2 + y2 * a,
+            y2 - x2 * a
+        ),
+        (
+            x2,
+            y2
+        )
+    ]
+
+def vectorAngle(ux, uy, vx, vy):
+    sign = -1 if (ux * vy - uy * vx < 0) else 1
+
+    dot = min(1, max(-1, ux * vx + uy * vy))
+
+    return sign * math.acos(dot)
+
+def getArcCenter(
+    px,
+    py,
+    cx,
+    cy,
+    rx,
+    ry,
+    largeArcFlag,
+    sweepFlag,
+    sinphi,
+    cosphi,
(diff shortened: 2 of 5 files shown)


mathandy:
Hi @njhurst, thanks for the PR.  
"converting arcs to line segments" -- when does this happen?  
Look like your PR changes disvg, what's the reason for that?

njhurst:
Golly, it's been a while, let me see if I remember :)
I think the goal was to represent arcs with cubic beziers, so that you're creating say 4 for a 0.4% error, rather than the 100 or so line segments.  This makes things like intersection and arc length much cheaper, as well as the corresponding render time.  I recall it made one of your tests take a 100ms rather than a minute to run to completion.

The change to disvg was factoring out the part that writes svg paths to files from the part that displays so that I can use your library directly in jupyter-python as an inline element.  I should have perhaps made a separate diff for this, but I'm naughty.  Sorry.

tatarize:
There are other speedups that can be done for the linearization. For @abey79's spearheaded a project on `svgelements` to use np arrays to much more quickly convert arcs to points. If you're hyper concerned with speed you'd likely already have numpy.

I think his benchmarks even for the svgpathtools package ( see svgelements/issues#52 ). Had the parallelized version going like 2.7x faster.

---

However, for simply changing Arcs into Beziers, I actually recommend this. I do it myself in places. Arcs are sometimes kinda rude segments and a lot weird.

I think #103 had this and regebro/svg.path#65 has an implementation of some of my code into something more like svg.path-like code. (my `svgelements` project needs arbitrary matrices applied to arcs so the parameterization is completely changed.) I think regebro/svg.path#59 at the bottom I did the actual legwork to convert the function into one that would work in `svgpathtools` as well as `svg.path`. 

It seems like this might be best done by in-place insertion of the bezier curves in place of the arcs. So you have a `Path` class and you you have some graphics library or something that does not include Arcs or does include Arc but only circular arcs but can handle bezier curves. You would then do `path.approximate_arcs_with_cubics()` or `path.approximate_arcs_with_quads()`. These functions would take an `error` parameter, even if it's not implemented. Then regardless of what your path was you'd know it doesn't have any Arcs in it. And you didn't have to linearize the entire shebang and need to deal with 100 points and higher error.

njhurst:
Yeah, there are lots of bigger versions of my patch, I just needed something specific and figured others might benefit.  I'm definitely interested in a fully vectorized version with numpy, but as I've already written one complete library that does these things I didn't want to get drawn in too deeply :-D  

I also have a complete port of libspiro to run on python+svgpathtools, if people are interested in that I can probably work out how to submit a patch.

tatarize:
I'm going to write up an easier patch for this bit of code. That should get accepted more easily since I can generally do the work to prove it it harmless. And it's fairly close already, and I could likely use the code it I'll have to write myself for my project since I already have these functions. And my last pull got added so I've got a place to fork my stuff.

I'm generally getting a more vectorized version but soft requirement on numpy (meerk40t/svgelements#61) . `svgelements` focuses much more directly on doing robust spec-correct parsing of the SVG throughout. So it has a bunch more things in it. And doesn't focus much on the math heavy stuff here. In fact I generally abduct math from this library when it's needed. But, my Arc code is generally parameterized differently so code to get rid or Arcs would help with compatibility.

tatarize:
@njhurst Can you check #130 and tell me if there's something your pull request here could do that that my pull there can't or doesn't. I see that your PR had some special side ability to use the Mortenson approximation for quarter curves. And while that's actually a better than the native 4/3Tan method it does not generalize. Did yours have some functionality I missed?

I mean this is clean enough to get an easy include and those features might need to get a different PR. But, it's worth having you check it out.

abey79:
I haven't looked in the details of this PR but, if Arcs are to be linearised, I'd like to raise awareness for #126.

tatarize:
Having reread it I think he wants arcs replaced with beziers, because lines are too slow. And it's *much* easier replace arcs with beziers if you have the ability to use them and keep using them instead of arcs rather exclusively. This is both true and quite useful for a variety of things. For example transforming things is really hard unless they are made entirely of points like CubicBeziers. You could use the #130 code to make a kind of good enough transformation matrix without doing a radical re-parameterization of arcs like I had to do for `svgelements`.

tatarize:
@njhurst This should maybe be closed. #130 should have this feature covered without issue.

mathandy:
@njhurst I'm closing this per @tatarize's recommendation. If there is reason (e.g. #130 doesn't cover your needs/concerns), please feel free to re-open.

njhurst:
Sure, fine by me!

PR #130, opened 2020-11-22
Approximate Arcs With Beziers

### LINKED ISSUE #103: Convert arc to cubic bezier

tatarize:
Closes #103
Replaces #116 

Adds in the ability to replace all Arcs within a Path with either Cubic or Quadratic bezier curves. Adds in the ability to perform this action on on the Arc line segments themselves.

### DIFF
--- a/svgpathtools/path.py
+++ b/svgpathtools/path.py
@@ -4,7 +4,7 @@ Arc."""
 
 # External dependencies
 from __future__ import division, absolute_import, print_function
-from math import sqrt, cos, sin, acos, asin, degrees, radians, log, pi, ceil
+from math import sqrt, cos, sin, tan, acos, asin, degrees, radians, log, pi, ceil
 from cmath import exp, sqrt as csqrt, phase
 import re
 try:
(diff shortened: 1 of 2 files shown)


svgpathtools/path.py
```diff
@@ -2906,6 +2986,32 @@ def is_contained_by(self, other):
...
+        for s in range(len(self)-1, -1, -1):
+            segment = self[s]
+            if not isinstance(segment, Arc):
+                continue
+            arc_required = int(ceil(abs(segment.delta) / sweep_limit))
+            self[s:s+1] = list(segment.as_cubic_curves(arc_required))
+
+    def approximate_arcs_with_quads(self, error=0.1):
+        """
+        Iterates through this path and replaces any Arcs with cubic bezier curves.
```
njhurst, line 3004: quadratic

njhurst (approved):

tatarize:
Added tests for the code. I'm kinda shocked how much better Cubic is than Quad. But I guess it makes a lot of sense.
@mathandy I'd be surprised if there was any objection to this code.

njhurst:
The arc length test was set with a really poor target precision so that it would run quickly, I couldn't find it any more, but you should update it to the same precision as the other cases with this change.  Cubic beziers converge as O(e^6), Quadratic as O(e^4), so cubics quickly dominate.  For geometric approximation it's even more stark, but I don't recall the convergence rates.

mathandy:
Thanks

Issue #125, opened 2020-11-12
colors are not taken into account in wsvg if path parameter is an instance of Path


piscvau:
Line 207 in paths2svg : 
if isinstance(paths, Path) or is_path_segment(paths):
        paths = [paths]
 if a path is submitted with a list of colors why embed path in a list again.

I would propose the following correction: replace line 207 with
if is_path_segment(paths) :

mathandy:
@piscvau I'm not following, can you give me an example?

E.g. I thought your were saying the following wouldn't work, but it does afaik
```
from svgpathtools import *
path = Path(Line(0,1+1j), Line(1+1j,2+1j))
disvg(path, 'r')
```

I'm closing this, but feel free to reopen if I've misunderstood.

PR #132, opened 2020-11-25
pull request for correction of issue #125 and add of split_extremum_y function to Arc, CubicBezier and QuadraticBezier

piscvau:
Test file for split_extremum_y is included

### DIFF
--- a/svgpathtools/path.py
+++ b/svgpathtools/path.py
@@ -6,6 +6,8 @@ Arc."""
 from __future__ import division, absolute_import, print_function
 from math import sqrt, cos, sin, acos, asin, degrees, radians, log, pi, ceil
 from cmath import exp, sqrt as csqrt, phase
+from scipy.optimize import newton
+from copy import deepcopy
 import re
 try:
     from collections.abc import MutableSequence  # noqa
@@ -1019,6 +1021,36 @@ class QuadraticBezier(object):
         bpoints1, bpoints2 = split_bezier(self.bpoints(), t)
         return QuadraticBezier(*bpoints1), QuadraticBezier(*bpoints2)
 
+    def split_extremum_y(self, t_tol=0.001, fprime_tol=0.001):
+        """
+        Split self in 2 curves at its extremum Point.
+
+        Args
+        ----
+            t_tol:      tolerance under which t1 and t2 are considered the same
+                        (default 0.001)
+            fprime_tol: tolerance under which fprime(t) is considered zero
+                        (default 0.001)
+
+        Returns
+        -------
+            c1, c2:     two segments, whose union is this segment and which
+                        join at the extremum point of self. If the extremum is
+                        the start, end, or if the curve is a line or a point,
+                        c1 is a copy of self and c2 is None.
+
+        """
+        try:
+            def fprime(t): return self.derivative(t).imag
+            s = newton(fprime, 0.5, tol=fprime_tol, maxiter=1000)
+            if s <= t_tol or 1 - s <= t_tol:
+                raise Exception()
+            c1, c2 = self.split(s)
+        except BaseException:
+            c1, c2 = deepcopy(self), None
+
+        return c1, c2
+    
     def cropped(self, t0, t1):
         """returns a cropped copy of this segment which starts at
         self.point(t0) and ends at self.point(t1)."""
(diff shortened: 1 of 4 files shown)


test/tests_split_extremum_y.py
```diff
@@ -0,0 +1,262 @@
...
+Documentation
+=============
+
+Usage
+=====
+
+
+@author: Odile Troulet-Lambert
+@copyright: Odile Troulet-Lambert
+@license: GPL
```
tatarize, line 45: GPL is significantly more restrictive than the MIT license the project is under.

svgpathtools/path.py
```diff
@@ -6,6 +6,8 @@
 from __future__ import division, absolute_import, print_function
 from math import sqrt, cos, sin, acos, asin, degrees, radians, log, pi, ceil
 from cmath import exp, sqrt as csqrt, phase
+from scipy.optimize import newton
```
tatarize, line 9: Why does this need `newton` from a new and extra, hard dependency? The code in the `bbox()` function already does this. Find the extreme point in both x and y to figure the bounding box. 
````python
n = self.start[0] - self.control[0]
d = self.start[0] - 2 * self.control[0] + self.end[0]
if d != 0:
     t = n / d
else:
    t = 0.5
````

Then split(t)

tatarize:
Also, I'm curious why this is needed. Is the desire here to rationalize the bezier curves? If that's needed why not make a path function called rationalize bezier curves? You can then solve for any inflection point within the bezier curve and break the curve at those points replacing the curves in the path in place. It would work a bit like #130 but rather than replacing arcs we'd be replacing irrational beziers with series of of rational ones. So, for example, the bounding box of all the bezier curves would be the extrema of the start and end points.

svgpathtools/path.py
piscvau: No problem this can be changed. I am unfamiliar with pull request how should I do it?

piscvau:
The need is for a knitting application where curves must be splitted in their extremum y to be knitted.

svgpathtools/path.py
piscvau: As i am completely not expert in this package I might have not seen that the bbox function could be used.

piscvau: However the code you propose above does not apply to arc

piscvau: for the arc, the function point_to_t does not work for arcs with rotation != 0. so I do not see how it is possible to go from the extremum - coordinates to the t to be able to split;  but again I am not an expert/

tatarize:
Hm. You could just convert those arcs to bezier curves using #130. Then it would but also you can calculate the extrema arc bit with the same code used to calculate the bbox there too. At least in theory. Though it would really be the pretty easy to split those too. So if you converted everything to a bunch of rational bezier curves would that suffice? Like nothing has a position either right or left or above or below the end points and And any inflections occur at the nodes?

piscvau:
why would this be any better than using the functions proposed in the pull request which are pretty straight forward?

tatarize:
Scipy is as a requirement tends to make things depend on more and more things. Anybody who uses svgpathtools would now also require scipy. 

> Prerequisites
> numpy
> svgwrite

This is the set of actual requirements declared in the readme. For example quad is used elsewhere but isn't a requirement since it falls back to other methods when and if that requirement isn't met. It would be somewhat acceptable to make a soft requirement so folks with scipy installed would get this functionality and those without would get some other functionality. However, here those aren't very hard functions to write. You solve for the extremas and split there. I'm pretty sure the math without scipy would be really quick anyway, and you'd tend to need fallback code, or maybe to throw NotImplementedError if they the soft requirement.

piscvau:
I clearly do not have the math expertise required to implement this other functionality you are talking about. Neither do I have the time, also considering that for my application needs, the full implementation of Arc intersection is a requirement not yet fulllfilled by the package.
 I propose to move the import of scipy within the split_extremum_function and in case the package is not installed to raise an exception. This I can easily do!......
Let us wait for the opinion of the owner of the repository.

tatarize:
If you ran the code to convert the arcs to bezier curves you'd actually be able to calculate the intersections and thus solve that needed data.

PR #135, opened 2020-12-09
Process path within an explicit SVG namespace

guillett:
Hi, first thanks a lot, your package is really fast and useful.

I am using an @evil-mad tool to generate paths from text and `path` nodes are generated with an explicit SVG namespace. (cf test file).

I only fixed the path collection as I am not aware of performance implication but I can all of the `getElementsByTagName(a)` by `getElementsByTagNameNS('http://www.w3.org/2000/svg', a)` if you prefer.


mathandy:
Thank you for the pull-request, however, I fear this would break some code.  Once the #179 regression tests are implemented, it will be easier to check for these things.  Feel free to file an issue or re-open this pull-request if there's still an issue.

Issue #138, opened 2021-01-15
Python IO streams (io.IOBase) support for SVG data reading and writing


pinkeen:
Right now relevant functions support only a file path string arguments which is quite limiting.

Certain use-cases require that data is piped via **stdio** and it would be very convenient to provide this option for  **CLI** tools using _svgpathtools_. There are some cases where it's preferable to juggle raw SVG data strings in memory. Currently the above can only be achieved by using temporary disk files - this is quite inelegant and inefficient.

It should be fairly easy to implement this transparently by checking if argument is of type `str` or `io.TextIOBase`. This would also allow using `io.StringIO` for juggling in-memory strings or custom streams.

mathandy:
Closing as this is I believe this is fixed by #176

PR #139, opened 2021-01-15
Implement IO streams support for SVG data I/O

pinkeen:
Resolves https://github.com/mathandy/svgpathtools/issues/138

### DIFF
--- a/svgpathtools/paths2svg.py
+++ b/svgpathtools/paths2svg.py
@@ -7,6 +7,7 @@ See also the document.py submodule.
 from __future__ import division, absolute_import, print_function
 from math import ceil
 from os import path as os_path, makedirs
+from io import IOBase
 from tempfile import gettempdir
 from xml.dom.minidom import parse as md_xml_parse
 from svgwrite import Drawing, text as txt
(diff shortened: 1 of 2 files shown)


tatarize:
I believe the internal code is using a ElementTree. I think that that actually already accepts either a string or a stream. I'll double check. Also, I would guess that duck casting in python is strongly preferred. I mean if I have something that you can open and read data from, that's a stream even if it's not strictly speaking an IOBase descendent. Looks like a duck, quacks like a duck...

tatarize:
```python
def parse(source, parser=None):
    """Parse XML document into element tree.

    *source* is a filename or file object containing XML data,
    *parser* is an optional parser instance defaulting to XMLParser.

    Return an ElementTree instance.

    """
    tree = ElementTree()
    tree.parse(source, parser)
    return tree
```

It appears like the code here is limited in the Document class. It believed the only input was filename. The better correction therefore is to save `self.original_filename` if the input type is a string, else None. Then it will save this input stream filename type as `unknown.svg` if you try to export it at the end. And whatever input you have there should be fed directly in the ElementTree parse. I'd check whether it's a string or not rather than IOBase.

After all, the ElementTree isn't limited to IOBase objects either:

```
        close_source = False
        if not hasattr(source, "read"):
            source = open(source, "rb")
            close_source = True
```

It checks whether than source has a read component and then goes ahead and opens it accordingly. If I wrote my own class that had a read and open, I could hand it to ElementTree but not to your code. Your code would decide since it isn't a IOBase it must be a str, even if it wasn't a string.

mathandy:
Thanks @tatarize and @pinkeen.  @pinkeen, please address @tatarize's concerns.

mathandy:
Closing due to #176 being merged in

PR #143, opened 2021-02-24
Correcting a few functions not handling Path having empty _segments

wesbz:


tatarize:
This is solid.

laroo:
Great! This MR fixed my SVG import errors

mathandy:
Thanks @wesbz and @tatarize and @laroo !

PR #149, opened 2021-05-03
add support for units and handle mindim=None correctly.

njhurst:
This PR adds support for real-world units in generated svg files (useful for generating CNC inputs, for example).  It also fixes a bug where mindim=None crashes, rather than unconstraining the viewport size as you would expect.


mathandy:
Thanks @njhurst this looks good.  One request though, please move the new argument `baseunit` to be the last argument taken in.  That way we won't mess up anyone's positional arguments when I merge this.

PR #155, opened 2021-09-22
Fix implementation of points()

saraedum:
so the argument `ts` is understood as a list of points and not as a parameter to `poly()`.


mathandy:
Thanks @saraedum !

PR #161, opened 2021-12-07
add support for rounded rectangles

CatherineH:
parse the rx and ry attributes of svg rectangles and add them as arcs to the path.


mathandy:
Thanks @CatherineH!

PR #168, opened 2022-01-30
Added strings as input

user123-source:


mathandy:
Closing as this is now possible thanks to #176.  `io.StringIO` are now supported by both the `Document` class and `svg2paths()` function.  There is also now a `Document.from_svg_string()` constructor.

PR #191, opened 2022-12-04
Fastfail Intersection

tatarize:
Intended to parallel and complement #188. This better optimizes the intersections code to be more correctly measured against PR 188 in timed tests. Adding in a rather hefty acceleration structure should be measured against better fast-failed implementations of the intersection code.


mathandy:
@tatarize this is great!  What an improvement.  Your new test is really a great tool too.  I'm merging this.  Very nice. Thanks and hope you're well!

PR #217, opened 2023-11-19
Optimize intersection feature for Quadratic Bezier curves

rBazan98:
This approach determines intersections by directly solving the parametric equations for Cartesian coordinates associated with each Bezier curve instead of finding intersections by interpolation.

Notation used is as follows:

- 'xn' and 'yn' represents coordinates of each control points 0, 1, and 2.
- Parametric forms of coordinates used: X = At^2 + Bt + x0 and Y = Ct^2 + Dt + y0.
- Equation at^4 + bt^3 + ct^2 + d = 0 identifies intersections as roots.


mathandy:

PR #218, opened 2023-11-24
Add error message for invalid Path point range

AndreVallestero:
As described on the README, "these parameterizations are defined over the domain 0 <= t <= 1". Currently, the user gets "Something went wrong" which is not very helpful.


mathandy:

PR #222, opened 2024-06-21
Bugfix for parsing rounded rect instances

kennyweiss:
This PR contains a minor bugfix for parsing rounded rectangles.

When calling `rect2pathd` through the `Document` class, the passed in `Element` instance did not directly expose its `rx` and `ry` attributes, which can be accessed through the `keys()` method.

I added unit tests for this case and verified that everything continues to work when calling `rect2pathd` with a dictionary and with the `svg2paths` function on an SVG document.


mathandy:
Great PRs @kennyweiss  Thank you!

Issue #223, opened 2024-07-02
Reading in dpaths introduces a lot of floating point approximation noise


ohnorobo:
Example

```
import svgpathtools
d = "M70.63,10.42c0.11,0.33-0.89,2.09-1.54,2.45c-4.95,2.73-17.52,7.24-39.46,11.04"  # path from KanjiVG
p = svgpathtools.parse_path(d)
# some kind of simple path modification not yet implemented
print(p.d())
```

outputs

```
M 70.63,10.42 C 70.74,10.75 69.74,12.51 69.08999999999999,12.870000000000001 C 64.13999999999999,15.600000000000001 51.56999999999999,20.11 29.62999999999999,23.91
```

I understand the performance gain from using floats, and I don't mind a bit of jitter, but I would like to keep my svg files somewhat clean if possible, also it really increases the text size. 

Is it possible to allow something like rounding to n decimal places in `.d()`? (I would be happy to take a crack at implementing this.) Or maybe there's another good way to solve this I haven't understood?

ohnorobo:
This is fixed, see https://github.com/mathandy/svgpathtools/pull/235

PR #225, opened 2024-07-04
Format floating point numbers in output d-paths

ohnorobo:
Format the floating point numbers output by `Path.d` to avoid a bunch of .x0000000000001 and .x999999999999 noise. Fixes https://github.com/mathandy/svgpathtools/issues/223.

Changed behavior:

- Numbers with small errors introduced by their floating point representation are output correctly. ex `12.870000000000001` -> `12.87`, `64.13999999999999` -> `64.14`

- Whole numbers which were emitted with a trailing 0 now have no trailing 0. ex: `2.0` -> `2`. This required some changes in the existing test cases. The output now more closely matches the way numbers are represented in the SVG spec examples. It's easiest to see this in the changes to `test.test_generation.TestGeneration.test_path_parsing`.

- Some small floating point numbers now emit fewer significant digits. ex:  `206.07112` -> `206.071` in `test.test_groups.TestGroups.test_add_group`. This is chosen by python's idea of which representation is the 'best' as explained [here](https://docs.python.org/3/tutorial/floatingpoint.html)

This change also removes the need for special python2/3 casing in `test.test_generation.TestGeneration.test_path_parsing` since the behavior of `g` is the same in both. It may be possible to further simplify this test.


faisalakhlaq:
Seems like there is no one merging the PRs?

mathandy:
Thanks for the PR @ohnorobo, I think it's a tempting idea, I certainly dislike the representation error and am open to solutions. That said, I think the solution should work for 15 digit decimals -- I think all representable 15 digit decimals can be converted to IEEE doubles and back without error ([wikipedia reference](https://en.wikipedia.org/wiki/Double-precision_floating-point_format)).

This solutions seems pretty aggressive, e.g.
```
In [1]: 'M {}'.format(64.13999)
Out[1]: 'M 64.13999'

In [2]: 'M {:g}'.format(64.13999)  # using g rounds pretty aggressively
Out[2]: 'M 64.14'
```
`svgpathtools` is intended for mathematical analysis of data stored as svg paths. For this reason, I don't think it would suit the primary customer base to round so aggressively.

If you have a good argument or other reason, please feel free to re-open this PR.

Issue #232, opened 2025-02-12
Enhancement to svgpathtools


NGExplorer:
While testing with NXP IMXRT device we find some issue in svgpath tools.

- Some GPU does not have 'arc' so we can use cubic bizen for such case.
- When GPU does not support 'arc' command Circle can be rendered using 2 arcs
- For polyline we should not add end path(z) command all times.
- When some parameters are not specified, assume it to zero as per SVGT12 specification

PR #233, opened 2025-02-12
Upstream svgpathtools improvement

NGExplorer:
Fixes [#232](https://github.com/mathandy/svgpathtools/issues/232)

Some GPU does not have 'arc' so we can use cubic bezier for such case.
When GPU does not support 'arc' command Circle can be rendered using 2 arcs

@mathandy Please review and let us know if any changes are required.

### DIFF
--- a/svgpathtools/svg_to_paths.py
+++ b/svgpathtools/svg_to_paths.py
@@ -31,8 +31,8 @@ def ellipse2pathd(ellipse):
     """converts the parameters from an ellipse or a circle to a string for a 
     Path object d-attribute"""
 
-    cx = ellipse.get('cx', 0)
-    cy = ellipse.get('cy', 0)
+    cx = (ellipse.get('cx', 0) or 0)
+    cy = (ellipse.get('cy', 0) or 0)
     rx = ellipse.get('rx', None)
     ry = ellipse.get('ry', None)
     r = ellipse.get('r', None)
@@ -40,8 +40,8 @@ def ellipse2pathd(ellipse):
     if r is not None:
         rx = ry = float(r)
     else:
-        rx = float(rx)
-        ry = float(ry)
+        rx = float(rx or 0)
+        ry = float(ry or 0)
 
     cx = float(cx)
     cy = float(cy)
@@ -66,7 +66,11 @@ def polyline2pathd(polyline, is_polygon=False):
     if isinstance(polyline, str):
         points = polyline
     else:
-        points = COORD_PAIR_TMPLT.findall(polyline.get('points', ''))
+        raw_points = polyline.get('points', '')
+        if not raw_points:
+            points = [(0,0)]
+        else:
+            points = COORD_PAIR_TMPLT.findall(raw_points)
 
     closed = (float(points[0][0]) == float(points[-1][0]) and
               float(points[0][1]) == float(points[-1][1]))
@@ -97,8 +101,8 @@ def rect2pathd(rect):
     
     The rectangle will start at the (x,y) coordinate specified by the 
     rectangle object and proceed counter-clockwise."""
-    x, y = float(rect.get('x', 0)), float(rect.get('y', 0))
-    w, h = float(rect.get('width', 0)), float(rect.get('height', 0))
+    x, y = float(rect.get('x', 0) or 0), float(rect.get('y', 0) or 0)
+    w, h = float(rect.get('width', 0) or 0), float(rect.get('height', 0) or 0)
     if 'rx' in rect or 'ry' in rect:
 
         # if only one, rx or ry, is present, use that value for both
--- a/svgpathtools/svg_to_paths.py
+++ b/svgpathtools/svg_to_paths.py
@@ -86,11 +86,7 @@ def polyline2pathd(polyline, is_polygon=False):
     if isinstance(polyline, str):
         points = polyline
     else:
-        raw_points = polyline.get('points', '')
-        if not raw_points:
-            points = [(0,0)]
-        else:
-            points = COORD_PAIR_TMPLT.findall(raw_points)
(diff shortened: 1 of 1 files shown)


nicusorcitu:
@mathandy can you please review and integrate this PR?
Thanks.

svgpathtools/svg_to_paths.py
```diff
@@ -60,39 +78,43 @@ def polyline2pathd(polyline, is_polygon=False):
...
-    d = 'M' + 'L'.join('{0} {1}'.format(x,y) for x,y in points)
+    d = 'M ' + ' L '.join('{0} {1}'.format(x,y) for x,y in points)
     if is_polygon or closed:
-        d += 'z'
+        d += ' z'
     return d
 
 
-def polygon2pathd(polyline):
+def polygon2pathd(polyline, is_polygon):
```
mathandy, line 102:
```suggestion
def polygon2pathd(polyline, is_polygon=True):
```
That way the change doesn't break anyone's code.

svgpathtools/svg_to_paths.py
```diff
@@ -60,39 +78,43 @@ def polyline2pathd(polyline, is_polygon=False):
     if isinstance(polyline, str):
         points = polyline
     else:
-        points = COORD_PAIR_TMPLT.findall(polyline.get('points', ''))
+        raw_points = polyline.get('points', '')
+        if not raw_points:
+            points = [(0,0)]
```
mathandy, line 83: `points = [(0,0)]`

This doesn't seem right, can you explain?

svgpathtools/svg_to_paths.py
```diff
@@ -31,27 +34,42 @@ def ellipse2pathd(ellipse):
...
-    d = ''
-    d += 'M' + str(cx - rx) + ',' + str(cy)
-    d += 'a' + str(rx) + ',' + str(ry) + ' 0 1,0 ' + str(2 * rx) + ',0'
-    d += 'a' + str(rx) + ',' + str(ry) + ' 0 1,0 ' + str(-2 * rx) + ',0'
+    PATH_KAPPA = 0.552284
+    rxKappa = rx * PATH_KAPPA;
+    ryKappa = ry * PATH_KAPPA;
 
-    return d + 'z'
+    #According to the SVG specification (https://lists.w3.org/Archives/Public/www-archive/2005May/att-0005/SVGT12_Main.pdf),
```
mathandy, line 56: I think this should be an option, we should let users continue using arcs if they want, and for now at least.

Let's make that the default unless there is a very good reason this new behavior should be the default.

We could
* add an argument like `use_cubics=False`
* we could also default to True if some environmental variable is set

mathandy:
> @mathandy can you please review and integrate this PR? Thanks.

Done! Sorry for the delay and thanks for the nice MR. I suggested some changes. Please keep in mind, I try to avoid breaking changes when possible and practical in this repo.

mathandy:
> > @mathandy can you please review and integrate this PR? Thanks.
> 
> Done! Sorry for the delay and thanks for the nice MR. I suggested some changes. Please keep in mind, I try to avoid breaking changes when possible and practical in this repo.

cc @NGExplorer

NGExplorer:
Hello @mathandy ,

Thank you for sharing improvements. We'll incorporate this suggestions and update you.

Best Regards,

Nilam

svgpathtools/svg_to_paths.py
NGExplorer: Hello @mathandy , We have incorporated this suggestion. We'll upstream our changes in this PR in 1-2 days.

svgpathtools/svg_to_paths.py
NGExplorer: Hello @mathandy , 

**Explanation:**
In polyline2pathd() Line [#L65](https://github.com/mathandy/svgpathtools/blob/master/svgpathtools/svg_to_paths.py#L65) closed variable computation don't validate if there are points.

e.g. in one of SVGT12 test vector following was generating exception
<polyline xml:id="polyline-02" fill="red" stroke="black" points="" />

As a result we used [(0,0)].

While more digging into SVGT12 specification we find that
Reference: https://www.w3.org/TR/SVG2/shapes.html#PolygonElement
    The initial value, (none), indicates that the polygon element is valid, but does not render.
As a result, we've modified our application to not call polyline2pathd() in such cases.

NGExplorer:
@mathandy ,

We have incorporated all of your suggestions, please review.

Best Regards,

Nilam

svgpathtools/svg_to_paths.py
```diff
@@ -121,7 +147,7 @@ def rect2pathd(rect):
     x2, y2 = x + w, y + h
     x3, y3 = x, y + h
 
-    d = ("M{} {} L {} {} L {} {} L {} {} z"
+    d = ("M {} {} L {} {} L {} {} L {} {} z"
```
mathandy, line 150: Does this change have an effect?

mathandy:
I need to update the CI. Working on it.

mathandy:
@NGExplorer CI is working again 🎉. I've also dropped support for python<3.8 (so we can now use type annotations, which I'm personally happy about). 

I made some additional requests, sorry I missed these issues the first time around.
Please rebase and make the changes and I think you'll be ready to go assuming CI passes.

I may need to run CI manually for you (due to github rules for first time contributors in a repo possibly), but I'll keep an eye out for updates and run it as soon as I see any.

Thanks again!

mathandy:
^ cc @NilamG

NGExplorer:
@mathandy 

Declining this PR vglite-tools-2.0 was for shape-only toolkit release.
Changes are incorporated into PR: https://github.com/mathandy/svgpathtools/pull/236
vglite-tools-2.1 include text,image support

PR #236, opened 2025-05-12
Upstream svgpathtools improvement - v2.1

NGExplorer:
Fixes https://github.com/mathandy/svgpathtools/issues/232

- Some GPU does not have 'arc' so we can use cubic bezier for such case.
- When GPU does not support 'arc' command Circle can be rendered using 2 arcs

Incorporated review comments mentioned in PR https://github.com/mathandy/svgpathtools/pull/233

@mathandy Please review and let us know if any changes are required.

### DIFF
--- a/svgpathtools/svg_to_paths.py
+++ b/svgpathtools/svg_to_paths.py
@@ -84,6 +84,9 @@ def polyline2pathd(polyline, is_polygon=False):
     else:
         points = COORD_PAIR_TMPLT.findall(polyline.get('points', ''))
 
+    if len(points) == 0:
+        return ''
+
     closed = (float(points[0][0]) == float(points[-1][0]) and
               float(points[0][1]) == float(points[-1][1]))
 
@@ -99,13 +102,13 @@ def polyline2pathd(polyline, is_polygon=False):
     return d
 
 
-def polygon2pathd(polyline):
+def polygon2pathd(polyline, is_polygon=True):
     """converts the string from a polygon points-attribute to a string 
     for a Path object d-attribute.
     Note:  For a polygon made from n points, the resulting path will be
     composed of n lines (even if some of these lines have length zero).
     """
-    return polyline2pathd(polyline, True)
+    return polyline2pathd(polyline, is_polygon)
 
 
 def rect2pathd(rect):
@@ -227,7 +230,7 @@ def svg2paths(svg_file_location,
     # path strings, add to list
     if convert_polygons_to_paths:
         pgons = [dom2dict(el) for el in doc.getElementsByTagName('polygon')]
-        d_strings += [polygon2pathd(pg) for pg in pgons]
+        d_strings += [polygon2pathd(pg, True) for pg in pgons]
         attribute_dictionary_list += pgons
 
     if convert_lines_to_paths:
--- a/svgpathtools/svg_to_paths.py
+++ b/svgpathtools/svg_to_paths.py
@@ -27,7 +27,7 @@ def path2pathd(path):
     return path.get('d', '')
 
 
-def ellipse2pathd(ellipse):
+def ellipse2pathd(ellipse, use_cubics=False):
     """converts the parameters from an ellipse or a circle to a string for a 
     Path object d-attribute"""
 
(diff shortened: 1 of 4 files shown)


mathandy:
Hi @NGExplorer, it looks like several of my requests / comments in #233 have not been taken care of yet. Please take a second look.

mathandy:
Also, would it be possible for you to add some test?

NGExplorer:
Hello @mathandy ,

On vglite-tools-2.1 branch we've incorporated following changes,
We rechecked your feedback on old PR #233 there were 4 review comments
1. Keep old behavior for 'arc' (keep use_cubics=False as default) to keep old toolkit behavior identical. ([link](https://github.com/mathandy/svgpathtools/pull/233/files/833a0a64c034cfe101196a47c6aab3683815bd9e))
    We have incorporated this change. ([fix](https://github.com/mathandy/svgpathtools/pull/236/commits/a48d21d5cc3965ab2cd890601126f2fb84f55a3a))

2. Use is_polygon variable in call to polyline2pathd
   We have incorporated this change. return polyline2pathd(polyline, is_polygon)
image

3. points = [(0,0)] ([link](https://github.com/mathandy/svgpathtools/pull/233#pullrequestreview-2804627612))   
    Failing vector [shapes-polygon-02-t.svg](https://www.w3.org/Graphics/SVG/Test/20080912/svggen/shapes-polygon-02-t.svg)
image
    Expected output: https://www.w3.org/Graphics/SVG/Test/20080912/htmlEmbedHarness/shapes-polygon-02-t.html
    We have modified our application(gpu-vglite-toolkit) to not invoke polyline2pathd(), when there are no points.

5. Remove spaces around VG commands ([link](https://github.com/mathandy/svgpathtools/pull/233/files/1396f997f1716c93d995e6677a28e608971eae23))
    We have removed its original patch() which introduced this issue.

Let us know if we missed something to improve svgpathtools.

svgpathtools/svg_to_paths.py
```diff
@@ -27,31 +30,54 @@ def path2pathd(path):
...
+        d += 'C' + str(cx + rxKappa) + ' ' + str(cy - ry) + ' ' + str(cx + rx) + ' ' + str(cy - ryKappa) + ' ' + str(cx + rx) + ' ' + str(cy)
+    else:
+        d = ''
+        d += 'M' + str(cx - rx) + ',' + str(cy)
+        d += 'a' + str(rx) + ',' + str(ry) + ' 0 1,0 ' + str(2 * rx) + ',0'
+        d += 'a' + str(rx) + ',' + str(ry) + ' 0 1,0 ' + str(-2 * rx) + ',0'
 
-    return d + 'z'
+    
+    return d + ' Z'
```
mathandy, line 80: Does this change have any effect?
(if no, let's remove it to avoid unnecessarily changing output)

svgpathtools/svg_to_paths.py
```diff
@@ -68,7 +94,7 @@ def polyline2pathd(polyline, is_polygon=False):
     # The `parse_path` call ignores redundant 'z' (closure) commands
     # e.g. `parse_path('M0 0L100 100Z') == parse_path('M0 0L100 100L0 0Z')`
     # This check ensures that an n-point polygon is converted to an n-Line path.
-    if is_polygon and closed:
+    if is_polygon or closed:
```
mathandy, line 97: What's the justification for this change?

I think this needs to be reverted, but if you have some reason for the change, we can talk about a solution

svgpathtools/svg_to_paths.py
```diff
@@ -77,22 +103,22 @@ def polyline2pathd(polyline, is_polygon=False):
...
 
 
 def rect2pathd(rect):
     """Converts an SVG-rect element to a Path d-string.
     
     The rectangle will start at the (x,y) coordinate specified by the 
     rectangle object and proceed counter-clockwise."""
-    x, y = float(rect.get('x', 0)), float(rect.get('y', 0))
-    w, h = float(rect.get('width', 0)), float(rect.get('height', 0))
+    x, y = float(rect.get('x', 0) or 0), float(rect.get('y', 0) or 0)
```
mathandy, line 120: What's the justification for these defaults?

svgpathtools/svg_to_paths.py
```diff
@@ -27,31 +30,54 @@ def path2pathd(path):
...
 
 
-def ellipse2pathd(ellipse):
+def ellipse2pathd(ellipse, use_cubics=False):
     """converts the parameters from an ellipse or a circle to a string for a 
     Path object d-attribute"""
 
-    cx = ellipse.get('cx', 0)
-    cy = ellipse.get('cy', 0)
+    cx = (ellipse.get('cx', 0) or 0)
```
mathandy, line 37: You've made several changes like this -- what's the justification? Is null a valid option here? If so, why change it to zero?

mathandy:
> Hi @mathandy Could you please review the changes incorporated by Nilam? If there is something else to be done just let us know. Thank you!

Done.
Also, please add unit tests for any changes to logic (assuming you have justification for them and want to keep them), as well as the new feature added.

svgpathtools/svg_to_paths.py
NGExplorer: No, this changes does not have side-effect. We will remove it.

svgpathtools/svg_to_paths.py
NGExplorer: In SVG, a <polygon> element is always rendered as a closed shape, regardless of whether the last point explicitly matches the first point. For example:

Test vector: [shapes-polygon-01-t.svg](https://www.w3.org/Graphics/SVG/Test/20080912/svggen/shapes-polygon-01-t.svg)

`<polygon xml:id="polygon-01" fill="none" stroke="#000000" points="59,45,95,63,108,105,82,139,39,140,11,107,19,65" />
`
In this case, the first and last points differ, so a 'closed' flag computed from point equality would be false. 

The original condition:

if is_polygon and closed:
    points.append(points[0])

fails to append the starting point because closed flag is false, even though 'is_polygon' is True. This can lead to incorrect processing in cases where polygon closure needs to be explicitly handled in the logic.

To account for the SVG specification and ensure consistent behaviour, this condition should be updated to:

if is_polygon or closed:
    points.append(points[0])

This change ensures that <polygon> elements are always treated as closed paths, aligning with the SVG specification.

svgpathtools/svg_to_paths.py
```diff
@@ -1,3 +1,6 @@
+#
+# Modified by NXP 2024
```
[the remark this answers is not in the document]
nicusorcitu: Hi @mathandy 
This is the modification copyright according to our NXP policy.

For the newly created files we have to add:
"Copyright 2025 NXP"

and just for those kind of files that already exist, but without copyright, we do not want to take any extra credit so this is why it says just:
"Modified by NXP 2025"

Please let me know if staying as is looks acceptable to you.

svgpathtools/svg_to_paths.py
NGExplorer: Reverted this change.

svgpathtools/svg_to_paths.py
mathandy: "This can lead to incorrect processing in cases where polygon closure needs to be explicitly handled in the logic."

@NGExplorer Can you give an example of this please?

NGExplorer: > "This can lead to incorrect processing in cases where polygon closure needs to be explicitly handled in the logic."
> 
> @NGExplorer Can you give an example of this please?

I got your point,
rect() and polygon() both uses polyline2pathd()

closure is implicit in polygon and is_polygon already sets to True in such cases.
Where is in case of polyline is_polygon is false.

In vector drawing close command can be specified using 'z' or Line (origin) command.
We are trying to force 'z' command where possible.

As a result if we compute closed by taking is_polygon in consideration, original condition can be retained.
Patch may look something like following, We will check this in our test-suite and update you by EOD.
```
     closed = (float(points[0][0]) == float(points[-1][0]) and
          float(points[0][1]) == float(points[-1][1]))
+    closed = closed or (is_polygon == True)

if is_polygon and closed:
    points.append(points[0])
    

+   if closed:
-   if is_polygon or closed:
        d += 'z'

```

NGExplorer:
@mathandy Please review and let us know if any changes are required.

svgpathtools/svg_to_paths.py
```diff
@@ -1,3 +1,6 @@
+#
+# Modified by NXP 2024, 2025
```
mathandy, line 2: Did you leave this attribution in by accident or is there some reason you really want it?

mathandy:
Looks great! I asked a question about the attribution, other than that, I think this is ready.

svgpathtools/svg_to_paths.py
nicusorcitu: hi @mathandy 
I left an explanation here: https://github.com/mathandy/svgpathtools/pull/236#discussion_r2092810815

Is this acceptable to stay in the code? It just says that we, NXP, modified some code on the offending file. It is NXP policy to enforce this upstream.

nicusorcitu: OK, we moved our modified copyright on the if block.
Thanks!

mathandy:
Thanks @nicusorcitu and @NGExplorer !

